PR #8, opened 2021-05-31
add solution

iedenysenko:


Vlad-Andr (approved):

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,32 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't add actor " + actor, e);
+        } finally {
+            if (session != null) {
+                session.close();
+            }
+        }
+        return actor;
```

src/main/java/mate/academy/hibernate/relations/model/Actor.java
```diff
@@ -1,8 +1,21 @@
...
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "actors")
 public class Actor implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String name;
+
```
Max-Pochepets, line 17:
```suggestion
```

Max-Pochepets (approved):

PR #12, opened 2021-06-01
first solution

andriiromaniuk:


src/main/java/mate/academy/hibernate/relations/model/Actor.java
```diff
@@ -1,8 +1,22 @@
...
+
+@Entity
+@Table(name = "actors")
 public class Actor implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String name;
+    @ManyToOne
+    @JoinColumn
```
DanyloIT, line 19: I am not convinced either we need this annotation or not..

DanyloIT (approved):
Seems Gucci

src/main/java/mate/academy/hibernate/relations/model/Actor.java
RomanDubovskyi, line 19: we don't need this annotation (JoinColumn), code will work just fine withour it

RomanDubovskyi (approved):

PR #23, opened 2021-06-01
first commit

dvdovynets:


Den4eg007 (approved):

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,32 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't save actor=" + actor + " to DB", e);
+        } finally {
+            if (session != null) {
+                session.close();
+            }
+        }
+        return actor;
```

Max-Pochepets (approved):

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
dvdovynets: Understood

PR #40, opened 2021-06-02
done

ritakisluhina:


src/main/java/mate/academy/hibernate/relations/util/HibernateUtil.java
```diff
@@ -1,9 +1,19 @@
...
 import org.hibernate.SessionFactory;
+import org.hibernate.cfg.Configuration;
 
 public class HibernateUtil {
+    private static SessionFactory instance = initSessionFactory();
+
+    private HibernateUtil() {
+    }
+
+    private static SessionFactory initSessionFactory() {
```
KononenkoKonstantin, line 12: the private method should be after public

KononenkoKonstantin (approved):

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,32 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't save actor to DB: " + actor, e);
+        } finally {
+            if (session != null) {
+                session.close();
+            }
+        }
+        return actor;
```

src/main/java/mate/academy/hibernate/relations/dao/impl/CountryDaoImpl.java
```diff
@@ -12,11 +15,32 @@ public CountryDaoImpl(SessionFactory sessionFactory) {
...
+        try {
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            session.save(country);
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't save country to DB", e);
```
Max-Pochepets, line 29:
```suggestion
            throw new DataProcessingException("Can't save country to DB: " + country, e);
```
be consistent

src/main/java/mate/academy/hibernate/relations/model/Actor.java
```diff
@@ -1,8 +1,18 @@
 package mate.academy.hibernate.relations.model;
 
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.ManyToOne;
+
+@Entity(name = "actors")
```
Max-Pochepets, line 9: this parameter only names the entity, if you want to specify table name - use `@Table(name = "actors")`

Max-Pochepets (changes requested):

src/main/java/mate/academy/hibernate/relations/model/Country.java
```diff
@@ -5,7 +5,7 @@
 import javax.persistence.GenerationType;
 import javax.persistence.Id;
 
-@Entity(name = "countries")
+@Entity
```
Max-Pochepets, line 8:
```suggestion
@Entity
@Table(name = "countries")
```
it's still better to call tables, which store info about multiple entities, in plural

Max-Pochepets (approved):

PR #70, opened 2021-08-02
Extends Files and implement classes

testYurka:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,26 @@
...
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).orElseThrow(
+            () -> new NoSuchElementException("Can't find actor for id: " + id));
```
docer95, line 24: There is no point in throwing NoSuchElementException with a simple method call method get() in Optional we will get the same error anyway

docer95 (approved):

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
KseniiaMakarova: @docer95 if we want to have our custom exception message - we will have to use `orElseThrow()`

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,14 @@
+<hibernate-configuration>
+    <session-factory>
+        <property name="hibernate.dialect">org.hibernate.dialect.MySQL5Dialect</property>
+        <property name="connection.url">jdbc:mysql://localhost:3306/ticket_app?serverTimezone=UTC</property>
+        <property name="connection.driver_class">com.mysql.cj.jdbc.Driver</property>
+        <property name="connection.username">root</property>
+        <property name="connection.password">1234</property>
+        <property name="show_sql">true</property>
+        <property name="hbm2ddl.auto">create-drop</property>
+        <mapping class="mate.academy.hibernate.relations.model.Movie"/>
```
KseniiaMakarova, line 10:
```suggestion
        <property name="hbm2ddl.auto">create-drop</property>
        
        <mapping class="mate.academy.hibernate.relations.model.Movie"/>
```

KseniiaMakarova (changes requested):
please make a new PR from a separate branch - not from master branch
the code is good

PR #73, opened 2021-08-05
JV-Hibernate-relations-hw-solution

olekskov:
Initial solution:
- Completed implementations of Model/Dao/Service layers
- Configured hibernate, added HibernateUtil
- Made PSVM work


RomanDubovskyi (approved):
all good!

PR #130, opened 2021-12-13
made it work finally

Andrey-Sidorkin:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,23 @@
...
     @Override
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).get();
```
nishchenkokateryna, line 21: Not sure that is a mistake, but maybe it would be better to throw RuntimeException()
And in other places too

nishchenkokateryna (approved):

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +16,34 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+        try {
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            session.save(actor);
+            transaction.commit();
+        } catch (HibernateException e) {
+            if (transaction != null) {
+                transaction.rollback();
+                throw new DataProcessingException(
+                        "Can't add actor '" + actor + "' to DB", e);
```
Sofasmile, line 30: let's move exception after if block
same in other dao

Sofasmile (changes requested):

Sofasmile (approved):

PR #142, opened 2021-12-20
Hibernate relations was created

Lione2009:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,26 @@
...
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).orElseThrow(() ->
+                new RuntimeException("Actor with id" + id + "not found"));
```
Oleksii-kho, line 24:
```suggestion
                new RuntimeException("Actor with id " + id + " not found"));
```

Oleksii-kho (changes requested):

Oleksii-kho (approved):

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +16,32 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
     }
 
     @Override
     public Optional<Actor> get(Long id) {
-        return null;
+        try (Session session = factory.openSession()) {
+            return Optional.ofNullable(session.get(Actor.class, id));
+        } catch (HibernateException exception) {
+            throw new DataProcessingException("Can't get actor by id=" + id, exception);
+        }
```
issugy, line 45: its not a mistake, but FYI: u won't get exception if there is no entity with this id
same for all get() methods below

issugy (approved):

PR #156, opened 2021-12-22
Movie, Actor and Country models were created. Dao and Service layers …

farfardesu:
…were implemented

### DIFF
--- a/src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
+++ b/src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
@@ -2,8 +2,11 @@ package mate.academy.hibernate.relations.dao.impl;
 
 import java.util.Optional;
 import mate.academy.hibernate.relations.dao.ActorDao;
+import mate.academy.hibernate.relations.exception.DataProcessingException;
 import mate.academy.hibernate.relations.model.Actor;
+import org.hibernate.Session;
 import org.hibernate.SessionFactory;
+import org.hibernate.Transaction;
 
 public class ActorDaoImpl extends AbstractDao implements ActorDao {
     public ActorDaoImpl(SessionFactory sessionFactory) {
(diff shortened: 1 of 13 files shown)


v20stepanenko (approved):

src/main/java/mate/academy/hibernate/relations/dao/impl/MovieDaoImpl.java
```diff
@@ -12,11 +15,33 @@ public MovieDaoImpl(SessionFactory sessionFactory) {
...
     }
 
     @Override
     public Optional<Movie> get(Long id) {
-        return null;
+        try (Session session = factory.openSession()) {
+            Movie movie = session.get(Movie.class, id);
+            return Optional.ofNullable(movie);
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get Country with id: " + id, e);
```
issugy, line 44:
```suggestion
            throw new DataProcessingException("Can't get Movie with id: " + id, e);
```

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,33 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
 
     @Override
     public Optional<Actor> get(Long id) {
-        return null;
+        try (Session session = factory.openSession()) {
+            Actor actor = session.get(Actor.class, id);
+            return Optional.ofNullable(actor);
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get Actor with id: " + id, e);
+        }
```
issugy, line 45: its not a mistake, but FYI: u won't get exception if there is no entity with this id

src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -2,10 +2,28 @@
...
+import javax.persistence.Table;
 
+@Entity
+@Table(name = "movies")
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+
```
issugy, line 22: empty line

issugy (changes requested):

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
farfardesu: should we add check like this instead of catch block?

if (Optional.ofNullable(actor).isEmpty()) {
     throw new RuntimeException("Can't get Actor with id: " + id);
}

issugy: we should check this value in the service layer

issugy (approved):

PR #185, opened 2022-03-23
Models of Country, Actor and Movie were implemented

nastyaprokhorenko:


src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,32 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
     public Actor add(Actor actor) {
-        return null;
+        Session session = null;
+        Transaction transaction = null;
+        try {
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            session.save(actor);
+            transaction.commit();
+        } catch (Exception e) {
```
karma-o, line 25: we can use this exception here and in other places too
```suggestion
        } catch (HibernateException e) {
```

karma-o (approved):

boroda4436 (approved):

PR #234, opened 2022-05-31
Hibernate relations have been implemented

Roman-Kachmaryk:


src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,33 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+        Transaction transaction = null;
+        try {
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            session.save(actor);
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+                throw new DataProcessingException("Can't add an actor " + actor, e);
```
Sofasmile, line 28: we should throw an exception in both cases when the transaction is not null and when the transaction is null, so let's move DataProcessingException after `if` statement

the same in other methods

src/main/java/mate/academy/hibernate/relations/model/Actor.java
```diff
@@ -1,8 +1,21 @@
...
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.ManyToOne;
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "actors")
+
```
Sofasmile, line 12: remove the empty line below the annotation

Sofasmile (changes requested):

Sofasmile (approved):

PR #246, opened 2022-06-13
solution 'hibernate relations' is done

DenysShl:


src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -2,21 +2,49 @@
...
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            session.save(actor);
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            log.error("Can`t save to DB actor: {}", actor, e);
+            throw new DataProcessingException("Can`t save to DB", e);
```
MaliukDaria, line 34:
```suggestion
            throw new DataProcessingException("Can`t save actor to DB " + actor, e);
```

same in other classes

MaliukDaria (changes requested):

MaliukDaria (approved):

PR #255, opened 2022-07-06
Implement task

gdpbalt:


src/main/java/mate/academy/hibernate/relations/model/Actor.java
```diff
@@ -1,8 +1,23 @@
...
+
+@Entity
+@Table(name = "actors")
 public class Actor implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String name;
+    @OneToOne
+    @JoinColumn(foreignKey = @ForeignKey(name = "FK_county_id"))
```
antonEEmus, line 20: Here should be a many to one relationship. many actors can be from one country.
I'm also not quite sure about JoinColumn and ForeignKey. i think you can manage without it, but maybe it's better to ask mentor. i, for example, didn't use it and everything was ok.

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,15 @@
...
+        <property name="connection.username">dev</property>
+        <property name="connection.password">Quaci_2sia0Me</property>
+        <property name="show_sql">true</property>
+        <property name="hbm2ddl.auto">create-drop</property>
+
+        <mapping class="mate.academy.hibernate.relations.model.Country" />
+        <mapping class="mate.academy.hibernate.relations.model.Actor" />
+        <mapping class="mate.academy.hibernate.relations.model.Movie" />
+    </session-factory>
+</hibernate-configuration>
```
antonEEmus, line 15: no empty line at the end of file

antonEEmus (changes requested):

src/main/resources/hibernate.cfg.xml
gdpbalt: Thank's for review
I will try to write more meanfull comments in the future :)

dSkarzh (approved):

PR #260, opened 2022-07-12
hw done

nazarbaklanov:
added dependencies to pom.xml
added implementation of all classes in package model
added implementation of all classes in package dao/impl and service/impl
added custom unchecked DataProcessingException
created hibernate.cfg.xml

BUILD SUCCESS

### DIFF
--- a/src/main/java/mate/academy/hibernate/relations/model/Movie.java
+++ b/src/main/java/mate/academy/hibernate/relations/model/Movie.java
@@ -1,11 +1,23 @@
 package mate.academy.hibernate.relations.model;
 
+import jakarta.persistence.Entity;
+import jakarta.persistence.FetchType;
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+import jakarta.persistence.OneToMany;
+import jakarta.persistence.Table;
 import java.util.ArrayList;
 import java.util.List;
 
+@Entity
+@Table(name = "movies")
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+    @OneToMany(fetch = FetchType.EAGER)
     private List<Actor> actors;
 
     public Movie() {
--- a/src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
+++ b/src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
@@ -1,16 +1,23 @@
 package mate.academy.hibernate.relations.service.impl;
 
+import mate.academy.hibernate.relations.dao.ActorDao;
 import mate.academy.hibernate.relations.model.Actor;
 import mate.academy.hibernate.relations.service.ActorService;
 
 public class ActorServiceImpl implements ActorService {
+    private final ActorDao actorDao;
+
+    public ActorServiceImpl(ActorDao actorDao) {
+        this.actorDao = actorDao;
+    }
+
     @Override
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).get();
     }
 }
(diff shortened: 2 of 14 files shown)


src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -1,11 +1,23 @@
...
+import jakarta.persistence.GenerationType;
+import jakarta.persistence.Id;
+import jakarta.persistence.OneToMany;
+import jakarta.persistence.Table;
 import java.util.ArrayList;
 import java.util.List;
 
+@Entity
+@Table(name = "movies")
 public class Movie implements Cloneable {
```
Cosmic-M, line 15: add to method toString:
+ ", actors list: " + actors

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,23 @@
...
     @Override
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).get();
```
Cosmic-M, line 21: Maybe it's better to throw exception from this layer. In this way you will cover two problems which can be.

Cosmic-M: orElseThrow(() -> new DataProcessingException("Cannot find any movie from DB with such id=" + id)) instead get()

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).orElseThrow(() ->
+                new DataProcessingException("Cannot find any actor from DB with such id: " + id));
```
KostyaNikiforov, line 23: This exception is supposed to be on the dao layer. This is not about data processing, but about the missing element. So is would be better to throw NSEE here

src/main/java/mate/academy/hibernate/relations/service/impl/CountryServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
     public Country add(Country country) {
-        return null;
+        return countryDao.add(country);
     }
 
     @Override
     public Country get(Long id) {
-        return null;
+        return countryDao.get(id).orElseThrow(() ->
+                new DataProcessingException("Cannot find any country from DB with such id: " + id));
```
KostyaNikiforov, line 23: same

src/main/java/mate/academy/hibernate/relations/util/HibernateUtil.java
```diff
@@ -1,9 +1,19 @@
 package mate.academy.hibernate.relations.util;
 
 import org.hibernate.SessionFactory;
+import org.hibernate.cfg.Configuration;
 
 public class HibernateUtil {
+    private static final SessionFactory instance = initSessionFactory();
+
+    public HibernateUtil() {
```
KostyaNikiforov, line 9: Why do you need to create an object of HibernateUtil class?

KostyaNikiforov (changes requested):

src/main/java/mate/academy/hibernate/relations/util/HibernateUtil.java
nazarbaklanov: Мы его используем для получения Session в main
( SessionFactory sessionFactory = HibernateUtil.getSessionFactory();)

nazarbaklanov: Перепутал, это в моделях нужен конструктор без параметров. Здесь он не нужен.

KostyaNikiforov (changes requested):
Fix build error and comment)

KostyaNikiforov (approved):

PR #266, opened 2022-07-17
models, services and dao were updated

mykola8426:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,24 @@
...
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id)
+                .orElseThrow(() -> new RuntimeException("Can't get actor by id: " + id));
```
antonEEmus, line 22: You can use NoSuchElementException here

antonEEmus (changes requested):

antonEEmus (approved):

liliia-ponomarenko (approved):
great 👍

PR #270, opened 2022-07-20
actor, country, movie models were implemented

KarinaTanasiichuk:


mykola8426 (approved):
Great solution!

liliia-ponomarenko (approved):
Awesome 😎

PR #276, opened 2022-07-20
-added my solution to task Hibernate-relations

kqlynero:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,23 @@
...
     @Override
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).get();
```
ChumakAndrey, line 21: I think its better to use orElseThrow there, to have NoSuchElement

src/main/java/mate/academy/hibernate/relations/util/HibernateUtil.java
```diff
@@ -1,9 +1,19 @@
 package mate.academy.hibernate.relations.util;
 
 import org.hibernate.SessionFactory;
+import org.hibernate.cfg.Configuration;
 
 public class HibernateUtil {
+    private static final SessionFactory sessionFactory = initSessionFactory();
+
+    public HibernateUtil() {
```
ChumakAndrey, line 9: you should use private constructor for singletone pattern

ChumakAndrey (approved):

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -2,21 +2,47 @@
...
 import java.util.Optional;
 import mate.academy.hibernate.relations.dao.ActorDao;
+import mate.academy.hibernate.relations.exception.DataProcessingException;
 import mate.academy.hibernate.relations.model.Actor;
+import org.hibernate.Session;
 import org.hibernate.SessionFactory;
+import org.hibernate.Transaction;
 
 public class ActorDaoImpl extends AbstractDao implements ActorDao {
+    private final SessionFactory sessionFactory = super.factory;
```
liliia-ponomarenko, line 12: Can you avoid using this field?

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,23 @@
...
     @Override
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).orElseThrow();
```
liliia-ponomarenko, line 21: Let’s  add an informative message

src/main/java/mate/academy/hibernate/relations/service/impl/CountryServiceImpl.java
```diff
@@ -1,16 +1,23 @@
...
     @Override
     public Country add(Country country) {
-        return null;
+        return countryDao.add(country);
     }
 
     @Override
     public Country get(Long id) {
-        return null;
+        return countryDao.get(id).orElseThrow();
```
liliia-ponomarenko, line 21: and here

src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,23 @@
...
     @Override
     public Movie add(Movie movie) {
-        return null;
+        return movieDao.add(movie);
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        return movieDao.get(id).orElseThrow();
```
liliia-ponomarenko, line 21: here too

liliia-ponomarenko (changes requested):

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
kqlynero: Sure, but we use this field in two methods, so I thought that I can use this here

PR #295, opened 2022-09-06
Hw-is-done

IrynaKhrustalova:


src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -1,11 +1,28 @@
...
 import java.util.List;
 
+@Entity
+@Table(name = "movies")
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+    @ManyToMany(fetch = FetchType.EAGER)
```
aaananas, line 22:
```suggestion
    @ManyToMany
```
fetch type EAGER is in general a bad practice, add fetchs instead to your queries

aaananas (approved):

PR #297, opened 2022-09-07
hw-hibernate-relations

RomanPlk:


src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -1,11 +1,28 @@
...
 import java.util.List;
 
+@Entity
+@Table(name = "movies")
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+    @ManyToMany(fetch = FetchType.EAGER)
```
aaananas, line 22:
```suggestion
    @ManyToMany
```
fetch type EAGER is a bad practice, lets add fetch join to queries

aaananas (changes requested):

src/main/java/mate/academy/hibernate/relations/model/Movie.java
RomanPlk: it'll be in the next topic, currently I don't know what _fetch join_ is

zachar-chorny (approved):

PR #303, opened 2022-09-17
Completed task

vitaliy1210:


src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
     @Override
     public Movie add(Movie movie) {
-        return null;
+        return movieDao.add(movie);
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        return movieDao.get(id).orElseThrow();
```
aaananas, line 23: if you use orElseThrow always throw an exception with message

src/main/java/mate/academy/hibernate/relations/exception/DataProcessingException.java
```diff
@@ -0,0 +1,8 @@
+package mate.academy.hibernate.relations.exception;
+
+public class DataProcessingException extends RuntimeException {
+
```
aaananas, line 4:
```suggestion
```

aaananas (changes requested):

aaananas (approved):

PR #326, opened 2022-09-27
Was implemented relations between different entities

valentyn-slutskyi:


valentyn-chepiha (approved):

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,24 @@
 package mate.academy.hibernate.relations.service.impl;
 
+import mate.academy.hibernate.relations.dao.ActorDao;
 import mate.academy.hibernate.relations.model.Actor;
 import mate.academy.hibernate.relations.service.ActorService;
 
 public class ActorServiceImpl implements ActorService {
+    private ActorDao actorDao;
```
dSkarzh, line 8: final

dSkarzh (approved):

PR #356, opened 2022-10-20
Hibernate relations

IgorPetrichenko:


src/main/java/mate/academy/hibernate/relations/service/impl/CountryServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
     @Override
     public Country add(Country country) {
-        return null;
+        return countryDao.add(country);
     }
 
     @Override
     public Country get(Long id) {
-        return null;
+        return countryDao.get(id).get();
```
aaananas, line 23: lets use orElseTrow and throw exception

aaananas (changes requested):

zachar-chorny (approved):

PR #366, opened 2022-11-16
implemented all tasks

nazariipertsak:


src/main/java/mate/academy/hibernate/relations/model/Actor.java
```diff
@@ -1,8 +1,20 @@
...
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "actors")
 public class Actor implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String name;
+    @OneToOne
```
YuriiHryniv, line 17: I think here we should use `@ManyToOne` instead because an actor can belong only to one single country and the country could have many actors. What do you think ? (but I am not 100% sure, let the mentor decide)

src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -2,10 +2,27 @@
...
+import javax.persistence.Table;
 
+@Entity
+@Table(name = "movies")
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+    @OneToMany(fetch = FetchType.EAGER)
```
YuriiHryniv, line 22: The same here. Movie can have many actors, and the actors could play a role in many films during their life (I used `@ManyToMany` instead, but I am not sure)

YuriiHryniv (approved):
LGTM

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,31 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+                session.close();
+            }
+        }
+        return actor;
     }
 
     @Override
     public Optional<Actor> get(Long id) {
-        return null;
+
```
hronograf, line 40: empty line

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,31 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+        try (Session session = factory.openSession()) {
+            return Optional.ofNullable(session.get(Actor.class, id));
+        }
```
hronograf, line 43: what if we have some exception here?
let's add catch block and throw our custom `DataProcessingException` with caught exception and some useful information like `id`
check this in all places

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,15 @@
...
+        <property name="connection.username">root</property>
+        <property name="connection.password">12345</property>
+        <property name="show_sql">true</property>
+        <property name="hbm2ddl.auto">create-drop</property>
+
+        <mapping class="mate.academy.hibernate.relations.model.Actor"/>
+        <mapping class="mate.academy.hibernate.relations.model.Country"/>
+        <mapping class="mate.academy.hibernate.relations.model.Movie"/>
+    </session-factory>
+</hibernate-configuration>
```
hronograf, line 15: files should end with empty line

hronograf (changes requested):
Good!

PR #383, opened 2022-11-30
relations implementation

alex-anofriev:


muzykaVP (approved):
LGTM!

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,35 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
     @Override
     public Optional<Actor> get(Long id) {
-        return null;
+        Actor actor;
+        try (Session session = factory.openSession()) {
+            actor = session.get(Actor.class, id);
+        } catch (RuntimeException e) {
+            throw new DataProcessingException("Can not find actor with id: " + id);
+        }
+        return Optional.ofNullable(actor);
```
ChumakAn, line 47:
```suggestion
        try (Session session = factory.openSession()) {
            actor = session.get(Actor.class, id);
            return Optional.ofNullable(actor);
        } catch (RuntimeException e) {
            throw new DataProcessingException("Can not find actor with id: " + id);
        }
```

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
     @Override
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).orElseThrow();
```
ChumakAn, line 23: you can use EntityNotFoundException and add an informational message

ChumakAn: same for all

ChumakAn (changes requested):
Good! Let's make it better

liliia-ponomarenko (approved):
Great 😊

PR #390, opened 2022-12-01
Models, DAO, services was implemented

muzykaVP:


src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -1,11 +1,23 @@
...
 import java.util.List;
 
+@Entity
+@Table(name = "movies")
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+    @OneToMany(fetch = FetchType.EAGER)
```
EdvardRoll, line 20: I think in this case, better to use ManyToMany annotation, because movie can have many actors and actors can act in many films

src/main/java/mate/academy/hibernate/relations/model/Actor.java
```diff
@@ -1,8 +1,21 @@
...
+import jakarta.persistence.Table;
+
+@Entity
+@Table(name = "actors")
 public class Actor implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String name;
+    @ManyToOne(fetch = FetchType.EAGER)
```
AndriiMatveev, line 18: It is not necessary to declare a strategy. In many to one by default this strategy

AndriiMatveev (approved):

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,31 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+        return actor;
     }
 
     @Override
     public Optional<Actor> get(Long id) {
-        return null;
+        Session session = factory.openSession();
+        Actor actor = session.get(Actor.class, id);
+        session.close();
+        return Optional.ofNullable(actor);
```
ChumakAn, line 43: let's use try with resources

ChumakAn: same for all

src/main/java/mate/academy/hibernate/relations/model/Actor.java
ChumakAn, line 18: let's remove fetch strategy and add annotation joinColumn () after that

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
     @Override
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).get();
```
ChumakAn, line 23: let's use orElseThrow with EntityNotFoundException and informational message

ChumakAn: same for all

ChumakAn (changes requested):
Good! Let's make it better

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,15 @@
...
+        <property name="connection.username">ulynhcyjrsdp1pit</property>
+        <property name="connection.password">0mnyNHAbgGS6DrQ7qchk</property>
+        <property name="show_sql">true</property>
+        <property name="hbm2ddl.auto">create-drop</property>
+
+        <mapping class ="mate.academy.hibernate.relations.model.Movie"/>
+        <mapping class ="mate.academy.hibernate.relations.model.Country"/>
+        <mapping class ="mate.academy.hibernate.relations.model.Actor"/>
+    </session-factory>
+</hibernate-configuration>
```
liliia-ponomarenko, line 15:
```suggestion
</hibernate-configuration>

```

src/main/java/mate/academy/hibernate/relations/Main.java
```diff
@@ -7,26 +7,31 @@
...
 
 public class Main {
     public static void main(String[] args) {
-        // use this session factory when you will initialize service instances
         SessionFactory sessionFactory = HibernateUtil.getSessionFactory();
 
         Country usa = new Country("USA");
-        CountryService countryService = null; // TODO: initialize this instance
+        CountryService countryService =
+                new CountryServiceImpl(sessionFactory);
```
liliia-ponomarenko, line 22: better pass the session factory only to the dao layer.

liliia-ponomarenko (approved):
Great 🔥

src/main/java/mate/academy/hibernate/relations/Main.java
muzykaVP: was doing so because i didn't want the dao layer appearance in main(presentation tier), but i already asked about it on Q\A and its ok in this case

PR #402, opened 2022-12-03
impl

sequencerr:


src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -1,11 +1,21 @@
...
 import java.util.List;
 
+@Entity
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+    @ManyToMany(fetch = FetchType.EAGER)
     private List<Actor> actors;
```
sequencerr, line 19: Seems my DBMS keeps names consistent 🥺
image

NSay0 (approved):
LGTM!

src/main/java/mate/academy/hibernate/relations/util/HibernateUtil.java
```diff
@@ -1,9 +1,16 @@
...
+    private static final SessionFactory INSTANCE =
+            new Configuration().configure().buildSessionFactory();
+
+    private HibernateUtil() {
+    }
+
     public static SessionFactory getSessionFactory() {
-        return null;
+        return INSTANCE;
     }
```
ChumakAn, line 15: let's add a method - initSessionFactory where you'll return the 
`new Configuration().configure().buildSessionFactory();`
so, you'll have a method for creating the configuration, therefore you can use it in 7 line and return the instance in the getSessionFactory as it's now.
Also, we can write an INSTANCE variable not in upperCase in this situation, but it's up to you

ChumakAn (changes requested):
Good! Let's make it better

sequencerr:
sorry, IntelliJ IDEA asked to commit "push anyway" second time that I missed

ChumakAn (approved):
LGTM

PR #411, opened 2022-12-08
Impl dao, service, add dependency in pom.xml

IvanAldokhin:


src/main/java/mate/academy/hibernate/relations/exception/DataProcessingException.java
```diff
@@ -0,0 +1,8 @@
+package mate.academy.hibernate.relations.exception;
+
+public class DataProcessingException extends RuntimeException {
+
```
Rommelua, line 4: line

Rommelua (approved):

PR #415, opened 2022-12-21
hibernate-relations-hw-solution

Pro1OOGamer:
1. added required hibernate dependencies
2. created unchecked DataProcessingException and throw it in the catch block on dao layer
3. completed implementation of all classes in the /model, /dao/impl and /service/impl package
4. created hibernate.cfg.xml file and implement HibernateUtil class
5. the code is tested in the main class with no errors

### DIFF
--- a/src/main/java/mate/academy/hibernate/relations/util/HibernateUtil.java
+++ b/src/main/java/mate/academy/hibernate/relations/util/HibernateUtil.java
@@ -1,9 +1,19 @@
 package mate.academy.hibernate.relations.util;
 
 import org.hibernate.SessionFactory;
+import org.hibernate.cfg.Configuration;
 
 public class HibernateUtil {
+    private static final SessionFactory instance = initSessionFactory();
+
+    public HibernateUtil() {
+    }
+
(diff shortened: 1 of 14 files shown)


src/main/java/mate/academy/hibernate/relations/util/HibernateUtil.java
```diff
@@ -1,9 +1,19 @@
 package mate.academy.hibernate.relations.util;
 
 import org.hibernate.SessionFactory;
+import org.hibernate.cfg.Configuration;
 
 public class HibernateUtil {
+    private static final SessionFactory instance = initSessionFactory();
+
+    public HibernateUtil() {
```
aaananas, line 9: why public?

aaananas (changes requested):

src/main/java/mate/academy/hibernate/relations/util/HibernateUtil.java
Pro1OOGamer: Because, the constructor generator in IDEA...(( sorry

aaananas (approved):

PR #419, opened 2023-01-11
implement actor/country/movie dao and service layer

Dmytro-Mk:


src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,36 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
     public Optional<Actor> get(Long id) {
-        return null;
+        Actor actor;
+        try (Session session = factory.openSession()) {
+            actor = session.get(Actor.class, id);
+        } catch (Exception e) {
+            throw new DataProcessingException(
+                    String.format("Can't get actor id = %s from DB", id), e);
+        }
+        return Optional.ofNullable(actor);
```
Nick97-git, line 48:
```suggestion
        try (Session session = factory.openSession()) {
            return Optional.ofNullable(actor);
        } catch (Exception e) {
            throw new DataProcessingException(
                    String.format("Can't get actor id = %s from DB", id), e);
        }
```
the same for other dao implementations

Nick97-git (changes requested):

Nick97-git (approved):

PR #436, opened 2023-02-11
hiber rel 1

Avassor3000:


src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,33 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
     }
 
     @Override
     public Optional<Actor> get(Long id) {
-        return null;
+        try (Session session = factory.openSession()) {
+            return Optional.ofNullable(session.get(Actor.class, id));
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get actor by id " + id, e);
+        }
```
prophet-dev-01, line 45: I consider, we don't need block catch in get method at the DAO level, because we return optional<Actor> and when we don't find an actor by this id, we return empty optional. You process this error on service level. This is only my opinion

prophet-dev-01 (approved):

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,33 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+        try {
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            session.save(actor);
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't save this actor", e);
```
Nick97-git, line 30:
```suggestion
            throw new DataProcessingException("Can't save actor: " + actor, e);
```
the same for add methods in other DAOs

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,33 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
 
     @Override
     public Actor add(Actor actor) {
-        return null;
+        Session session = null;
+        Transaction transaction = null;
+
```
Nick97-git, line 20: remove empty line, the same for add methods in other DAOs

Nick97-git (changes requested):

Nick97-git (approved):

PR #461, opened 2023-02-25
All tasks are done.

svitlana-tertyshna:


src/main/java/mate/academy/hibernate/relations/service/impl/CountryServiceImpl.java
```diff
@@ -1,16 +1,24 @@
...
     @Override
     public Country add(Country country) {
-        return null;
+        return countryDao.add(country);
     }
 
     @Override
     public Country get(Long id) {
-        return null;
+        return countryDao.get(id).orElseThrow(NoSuchElementException::new);
```
olekskov, line 22: It looks nice, no doubt, but we cannot skip adding proper information to exception thrown

src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,24 @@
...
     @Override
     public Movie add(Movie movie) {
-        return null;
+        return movieDao.add(movie);
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        return movieDao.get(id).orElseThrow(NoSuchElementException::new);
```
olekskov, line 22: same

olekskov (changes requested):

uliana-andrushchuk (approved):

PR #466, opened 2023-02-28
done

Richaa2:


src/main/java/mate/academy/hibernate/relations/model/Country.java
```diff
@@ -1,16 +1,24 @@
...
 
-    public Country() {
-    }
-
     public Country(String name) {
         this.name = name;
     }
 
+    public Country() {
+    }
```
Ivan95kos, line 20: For the future, it is better that the constructors go in ascending order, that is, the constructor without parameters should be the first

Ivan95kos (approved):

src/main/java/mate/academy/hibernate/relations/model/Country.java
Richaa2: Ok

PR #468, opened 2023-03-02
was solved

Andrii-Varlanovych:


src/main/java/mate/academy/hibernate/relations/service/impl/CountryServiceImpl.java
```diff
@@ -1,16 +1,23 @@
...
     @Override
     public Country add(Country country) {
-        return null;
+        return countryDao.add(country);
     }
 
     @Override
     public Country get(Long id) {
-        return null;
+        return countryDao.get(id).orElseThrow();
```
Ivan95kos, line 21: For the future, it is better to throw a custom exception or NoSuchElementException
```suggestion
        return countryDao.get(id).orElseThrow(() ->
                new NoSuchElementException("Can't get by id: " + id));
```

Ivan95kos (approved):

PR #472, opened 2023-03-07
done

diana-lomei:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,23 @@
...
     @Override
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).get();
```
uliana-andrushchuk, line 21: It's better to use .orElseThrow() method here

src/main/java/mate/academy/hibernate/relations/service/impl/CountryServiceImpl.java
```diff
@@ -1,16 +1,23 @@
...
     @Override
     public Country add(Country country) {
-        return null;
+        return countryDao.add(country);
     }
 
     @Override
     public Country get(Long id) {
-        return null;
+        return countryDao.get(id).get();
```
uliana-andrushchuk, line 21: It's better to use `.orElseThrow()` method here

uliana-andrushchuk (changes requested):

uliana-andrushchuk (approved):

PR #473, opened 2023-03-07
my hibernate-relations-hw is done

Vitalina-V:


src/main/java/mate/academy/hibernate/relations/util/HibernateUtil.java
```diff
@@ -1,9 +1,20 @@
...
+    private HibernateUtil() {
+    }
+
+    private static SessionFactory initSessionFactory() {
+        return new Configuration().configure().buildSessionFactory();
+    }
+
     public static SessionFactory getSessionFactory() {
-        return null;
+
```
Ivan95kos, line 17: For the future, remove the first empty line in all classes/interfaces/methods where it is present
```suggestion
```

Ivan95kos (approved):

PR #497, opened 2023-04-07
Added solution to the HW

ivbudenkov:


src/main/java/mate/academy/hibernate/relations/model/Actor.java
```diff
@@ -1,8 +1,20 @@
...
+import jakarta.persistence.Table;
+
+@Entity
+@Table(name = "actors")
 public class Actor implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String name;
+    @OneToOne
```
okuzan, line 17: wrong cardinality

okuzan (changes requested):

uliana-andrushchuk (approved):

PR #518, opened 2023-04-12
Relations

NadiaHuryk:


ArtemGrunin (approved):
Good job)

src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -61,6 +79,7 @@ public String toString() {
         return "Movie{"
                 + "id=" + id
                 + ", title='" + title + '\''
+                + ", actors='" + actors + '\''
```
petro-ushchuk, line 82: Bad idea to keep all actor's `toString` in movie `toString`. Imagine you whould have bidirectional relation instead of uni - that means you whould have list of movies in actors. So, you'll have endless loop of toString actors in movies in actors in movies and so on. So, if you really need to keep list of your actors in `toString`, there should be list of their ids to avoid circuit. But it is never needed, so you can just remove actors from here.

src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -2,10 +2,28 @@
...
+import javax.persistence.Id;
+import javax.persistence.JoinColumn;
+import javax.persistence.JoinTable;
+import javax.persistence.ManyToMany;
+import javax.persistence.Table;
 
+@Entity
+@Table(name = "movies")
 public class Movie implements Cloneable {
+
```
petro-ushchuk, line 18: remove all redundant empty lines

petro-ushchuk (changes requested):
Good work, but let's make it better

petro-ushchuk (approved):
Good work 🔥

PR #523, opened 2023-04-13
Implemented solution for HW.

RandomEastEcho:


NadiaHuryk (approved):

Andriy-dz (approved):
Very Good solution! Aproved!

PR #532, opened 2023-04-13
hw implementation

Deschna:


src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,34 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+        try {
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            session.persist(actor);
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't add actor " + actor);
```
nacenik, line 29: You should provide cause for your exception too. Fix it in all places

nacenik (changes requested):
Good solution, let's improve it :)

nacenik (approved):
LGTM! 🚀

PR #561, opened 2023-06-12
first version

OlehPryshchepa:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
     @Override
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).orElseThrow();
```
sZlishchev, line 23: Add an exception with an informative message. Check the following code for the same issue

src/main/java/mate/academy/hibernate/relations/exception/DataProcessingException.java
```diff
@@ -0,0 +1,7 @@
+package mate.academy.hibernate.relations.exception;
+
+public class DataProcessingException extends RuntimeException {
+    public DataProcessingException(String message) {
+        super(message);
```
sZlishchev, line 5: In my opinion, it's better to have a constructor with two parameters: message and exception object

rovein (changes requested):

maxline (approved):
Well done! Approved

PR #564, opened 2023-06-19
solved

MykytaKat:


src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -2,10 +2,22 @@
...
+import javax.persistence.Table;
 
+@Entity
+@Table(name = "movies")
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+    @OneToMany(fetch = FetchType.EAGER)
```
Tupadzaky, line 20: Why do you use OneToMany annotation here? 
One actor can perform in many movies, and one movie can have many actors, so in total it's many-to-many relationship

src/main/java/mate/academy/hibernate/relations/model/Actor.java
```diff
@@ -1,8 +1,20 @@
...
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "actors")
 public class Actor implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String name;
+    @ManyToOne
```
rovein, line 17: Let's also add mapping related to the foreign key (we need to provide it here, on the child entity side):
`@JoinColumn(name = "country_id")`

PR #592, opened 2023-07-05
addSolution

ArinaPolna:


VRuban373 (approved):
Good job:)

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,27 @@
...
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id)
+                .orElseThrow(() -> new DataProcessingException("Can't get actor by id" + id));
```
rovein, line 25:
```suggestion
                .orElseThrow(() -> new NoSuchElementException("Can't get actor by id" + id));
```
Please fix similar cases on the `service` layer (on `dao` layer the type of exception is correct).

src/main/java/mate/academy/hibernate/relations/exception/DataProcessingException.java
```diff
@@ -0,0 +1,7 @@
+package mate.academy.hibernate.relations.exception;
+
+public class DataProcessingException extends RuntimeException {
+    public DataProcessingException(String message) {
+        super(message);
+    }
```
rovein, line 6:
```suggestion
    public DataProcessingException(String message, Throwable cause) {
        super(message, cause);
    }

```

rovein (changes requested):
Good job! But let's improve your solution ;)

PR #598, opened 2023-07-05
Implemented task relations

Valdemar619:


src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,32 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            session.save(actor);
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't add actor " + actor + " to db ");
+        } finally {
```
ViktorBykh, line 30: Hereinafter, we should also add the cause of exception:
```suggestion
            throw new DataProcessingException("Can't add to db an actor with id: " + actor.getId(),  e);
        } finally {
```

src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -2,10 +2,22 @@
...
 
+@Entity
+@Table(name = "movies")
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+    @OneToMany(fetch = FetchType.EAGER)
     private List<Actor> actors;
```
ViktorBykh, line 21: It could be `@ManyToMany` relations
```suggestion
    @ManyToMany(fetch = FetchType.EAGER)
    @JoinTable(name = "movies_actors",
            joinColumns = @JoinColumn(name = "movie_id"),
            inverseJoinColumns = @JoinColumn(name = "actor_id"))
    private List<Actor> actors;   
```

ViktorBykh (approved):
Everything looks good) 
I have a few suggestions

src/main/java/mate/academy/hibernate/relations/model/Movie.java
sarakhmen: agree with @ViktorBykh , actor may act in many movies, so let's use many-to-many relation type

sarakhmen (changes requested):
Let's slightly improve your solution :)

maxline (approved):
Well done! Approved

PR #601, opened 2023-07-05
build 1.0

maksigors:


src/main/java/mate/academy/hibernate/relations/Main.java
```diff
@@ -1,32 +1,47 @@
...
+import mate.academy.hibernate.relations.service.impl.MovieServiceImpl;
 import mate.academy.hibernate.relations.util.HibernateUtil;
 import org.hibernate.SessionFactory;
 
 public class Main {
     public static void main(String[] args) {
-        // use this session factory when you will initialize service instances
         SessionFactory sessionFactory = HibernateUtil.getSessionFactory();
 
+        CountryDao countryDao = new CountryDaoImpl(sessionFactory);
```
DV7Art, line 26: i cant understand, why u need this Dao and in other

DV7Art (approved):
Alles gut..easy!

src/main/java/mate/academy/hibernate/relations/Main.java
maxline: @DV7Art I think it is not a problem, we inject dao via constructor instead of using new dao in the service classes. We had similar examples during learning SOLID topic.

maksigors: да, вже бачу, можна зробити бiльш ефективно

maxline (approved):
Well done from the first attempt! I hope you did this task by yourself)

PR #612, opened 2023-07-10
hibernate-relations-hw

VolodymyrDavidov:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,27 @@
...
+import mate.academy.hibernate.relations.dao.impl.ActorDaoImpl;
 import mate.academy.hibernate.relations.model.Actor;
 import mate.academy.hibernate.relations.service.ActorService;
+import org.hibernate.SessionFactory;
 
 public class ActorServiceImpl implements ActorService {
+    private ActorDao actorDao;
+
+    public ActorServiceImpl(SessionFactory sessionFactory) {
+        this.actorDao = new ActorDaoImpl(sessionFactory);
```
olekskov, line 14: It's not a good practice to initialize dependencies with new i constructor.
It's better to do it externally and pass as argument. So when we have framework to inject dependencies our code requires no changes

olekskov (approved):
Good job overall

PR #619, opened 2023-07-23
All done

oleksandrCod:


src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,24 @@
...
     @Override
     public Movie add(Movie movie) {
-        return null;
+        return movieDao.add(movie);
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        return movieDao.get(id).get();
```
olekskov, line 22: Directly calling get() on an Optional can lead to a NoSuchElementException if the Optional is empty. It's better to use orElseThrow() with a suitable exception.

src/main/java/mate/academy/hibernate/relations/service/impl/CountryServiceImpl.java
```diff
@@ -1,16 +1,24 @@
...
     @Override
     public Country add(Country country) {
-        return null;
+        return countryDao.add(country);
     }
 
     @Override
     public Country get(Long id) {
-        return null;
+        return countryDao.get(id).get();
```
olekskov, line 22: You are directly using get() on an Optional without checking if a value is present. This can lead to a NoSuchElementException if the Optional is empty. It's better to use the orElseThrow() method to throw an exception when the value is not present.

src/main/java/mate/academy/hibernate/relations/Main.java
```diff
@@ -1,32 +1,34 @@
...
-        // use this session factory when you will initialize service instances
-        SessionFactory sessionFactory = HibernateUtil.getSessionFactory();
-
+        final SessionFactory sessionFactory = HibernateUtil.getSessionFactory();
         Country usa = new Country("USA");
-        CountryService countryService = null; // TODO: initialize this instance
-        countryService.add(usa);
-
+        CountryService countryService = new CountryServiceImpl(new CountryDaoImpl(sessionFactory));
+        Country add = countryService.add(usa);
```
olekskov, line 24: why we have variable add? it doesn't seem to be used anywhere

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,38 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+        try {
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            session.save(actor);
+            transaction.commit();
+        } catch (DataProcessingException exception) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new RuntimeException("Can't save object to DB. " + actor);
```
olekskov, line 29: DataProcessingException is your custom exception, it cannot be caught during problems with DB. You should throw it:
```suggestion
        } catch (Exception exception) {
            if (transaction != null) {
                transaction.rollback();
            }
            throw new DataProcessingException("Can't save object to DB. " + actor);
```
Please recheck all similar occurences

olekskov (changes requested):

PR #630, opened 2023-08-02
hibernate relations for Movie, Actor and Country entities implemented…

TymurProkhorov:
…, all methods implemented as well.

### DIFF
--- a/src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
+++ b/src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
@@ -2,8 +2,11 @@ package mate.academy.hibernate.relations.dao.impl;
 
 import java.util.Optional;
 import mate.academy.hibernate.relations.dao.ActorDao;
+import mate.academy.hibernate.relations.exception.DataProcessingException;
 import mate.academy.hibernate.relations.model.Actor;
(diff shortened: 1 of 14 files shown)


Ivan95kos (changes requested):
Good job overall!

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,30 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+        Session session = null;
+        Transaction transaction;
+        try {
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            session.save(actor);
+            transaction.commit();
+        } catch (Exception e) {
+            throw new DataProcessingException("Can`t save actor: " + actor);
+        } finally {
```
olekskov, line 27: All operations except get/getAll should have transaction rollback to avoid data corruption
```suggestion
        } catch (Exception e) {
            if (transaction != null) {
                transaction.rollback();
            }
            throw new DataProcessingException("Can`t save actor: " + actor);
        } finally {
```
Please recheck all classes for same issue

olekskov (changes requested):

olekskov (approved):
Good job

PR #642, opened 2023-08-09
Hibernate Relations Solution

knyazkyylubomir:
implemented every model, implemented every dao, implemented every service


olekskov (approved):
Good job

PR #645, opened 2023-08-14
Implemented hibernate, Dao and Service sides

holsoni:


src/main/java/mate/academy/hibernate/relations/model/Actor.java
```diff
@@ -1,8 +1,21 @@
...
+import javax.persistence.Table;
+
+@Entity
+@Table(name = "actors")
 public class Actor implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String name;
+    @ManyToOne(fetch = FetchType.EAGER)
```
Denis-Balako, line 18: Many to one by default has eager fetch type

pom.xml
```diff
@@ -31,6 +31,16 @@
             <version>2.3.4</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.hibernate</groupId>
+            <artifactId>hibernate-core</artifactId>
+            <version>5.6.15.Final</version>
```
Denis-Balako, line 37: also, let's use the latest version "6.2.7.Final"

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,32 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
 
     @Override
     public Actor add(Actor actor) {
-        return null;
+        Session session = null;
+        Transaction transaction = null;
+        try {
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            session.save(actor);
```
Denis-Balako, line 23: It's not a good idea to use deprecated methods.
```suggestion
            session.persist(actor);
```

Denis-Balako (commented):
and a few little comments

pom.xml
holsoni: Why?

Vasul-Pankiv (approved):

PR #656, opened 2023-08-14
Hibernate relations solution

keda-vlad:


src/main/java/mate/academy/hibernate/relations/Main.java
```diff
@@ -16,17 +25,20 @@ public static void main(String[] args) {
         SessionFactory sessionFactory = HibernateUtil.getSessionFactory();
 
         Country usa = new Country("USA");
-        CountryService countryService = null; // TODO: initialize this instance
+        CountryDao countryDao = new CountryDaoImpl(sessionFactory);
```
Someboty, line 28: You can do it without creating dao variables

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,23 @@
...
     @Override
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).orElse(new Actor());
```
Someboty, line 21: Why do you return an empty actor, if you couldn't find one by ID? You should throw an exception in that case. Same in other services' implementations.

Someboty (changes requested):
Great job! Let's improve it a little.

src/main/java/mate/academy/hibernate/relations/Main.java
keda-vlad: What do u mean?

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
keda-vlad: between returning an empty object or throwing an exception I choose here first because I don`t have any requirements about this in README or Common mistake, it is almost the end of the program, and if it will return an empty object I just will see this in main. And tests go normally so I don`t think that it matters.

nacenik: But you should throw an exception. How can users understand that they don't have this row in the database?
fix it in all places

nacenik (changes requested):
Good solution. let's improve it :)

Vasul-Pankiv (approved):

PR #658, opened 2023-08-14
added solution

IhorKalaur:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
+import mate.academy.hibernate.relations.dao.ActorDao;
+import mate.academy.hibernate.relations.dao.impl.ActorDaoImpl;
 import mate.academy.hibernate.relations.model.Actor;
 import mate.academy.hibernate.relations.service.ActorService;
+import org.hibernate.SessionFactory;
 
 public class ActorServiceImpl implements ActorService {
+    private final ActorDao actorDao;
+
+    public ActorServiceImpl(SessionFactory factory) {
```
ArtemReizer, line 12: you need to pass dao to the arguments of the constructor. change it in all methods and in main()

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
     @Override
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).get();
```
ArtemReizer, line 23: what if you won't find any actor by id?

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,38 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+            }
+        }
+        return actor;
     }
 
     @Override
     public Optional<Actor> get(Long id) {
-        return null;
+        Session session = null;
+        try {
```
ArtemReizer, line 41: use try-with-resources in all get methods, because you don't have transactions in get methods

ArtemReizer (changes requested):
Good solution, let's improve it!))

ArtemReizer (approved):

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,30 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
 
     @Override
     public Actor add(Actor actor) {
-        return null;
+        Session session = null;
+        Transaction transaction = null;
+        try {
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            session.save(actor);
```
vovasalyha, line 23: It's better to use `persist`. You can google it why.

PR #666, opened 2023-08-15
added solution

Visqe:


src/main/java/mate/academy/hibernate/relations/model/Country.java
```diff
@@ -11,6 +19,8 @@ public Country(String name) {
         this.name = name;
     }
 
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
```
nacenik, line 23: it's not a field. It's a method. But you should add this annotation for the fields

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,15 @@
...
+        <property name="connection.username">root</property>
+        <property name="connection.password">password22</property>
+        <property name="show_sql">true</property>
+        <property name="hbm2ddl.auto">create-drop</property>
+
+        <mapping class="mate.academy.hibernate.relations.model.Actor"/>
+        <mapping class="mate.academy.hibernate.relations.model.Country"/>
+        <mapping class="mate.academy.hibernate.relations.model.Movie"/>
+    </session-factory>
+</hibernate-configuration>
```
nacenik, line 15: Add empty line at the end of the file

nacenik (changes requested):
Good, just small fixes

nacenik (changes requested):
Need to improve

vovasalyha (approved):

PR #673, opened 2023-08-15
hibernate-relations-hw completed

MarynaKushniruk:


src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
-        return null;
+        return movieDao.add(movie);
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+
+        return movieDao.get(id)
+                .orElseThrow(() -> new RuntimeException("Can't get movie by id " + id));
```
okuzan, line 23: change everywhere to NoSuchElement or EntityNotFound

okuzan (changes requested):

okuzan (changes requested):
to unblock on platform

olekskov (approved):
Good job

PR #680, opened 2023-08-22
Homework hibernate-relations.

UniyatYura:


src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
-        return null;
+
+        return movieDao.add(movie);
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        return movieDao.get(id).orElseThrow(() ->
+                new DataProcessingException("Can't get movie from DB with id " + id));
```
okuzan, line 23: thrown NoSuchElement or EntityNotFound instead

okuzan (changes requested):

okuzan (approved):

PR #706, opened 2023-09-21
added entity relations

RedArtelerist:


src/main/java/mate/academy/hibernate/relations/dao/impl/CountryDaoImpl.java
```diff
@@ -12,11 +15,30 @@ public CountryDaoImpl(SessionFactory sessionFactory) {
 
     @Override
     public Country add(Country country) {
-        return null;
+        Session session = null;
+        Transaction transaction = null;
+        try {
```
NazarSmith228, line 20: why not `try-with-resources`?

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,24 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+            transaction = session.beginTransaction();
+            session.persist(actor);
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't add actor to DB", e);
+        }
+        return actor;
```
NazarSmith228, line 29: it seems that implementations for `ActorDao`, `MovieDao` and `CountryDao` are identical. therefore it may be considered as a violation of DRY

you could create a method in abstract parent:
```
protected <T> void doAdd(T entity) {
}
```
and use that logic in this method. this way you would just call that method in each implementation so that the code isn't duplicated

however I'm not sure whether you're allowed to modify `AbstactDao` - you might ask the question in our Slack chat

if you're now allowed - feel free to skip this comment :)

src/main/java/mate/academy/hibernate/relations/model/Actor.java
```diff
@@ -1,8 +1,22 @@
...
+import jakarta.persistence.Table;
+
+@Entity
+@Table(name = "actors")
 public class Actor implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String name;
+    @ManyToOne
```
NazarSmith228, line 18: is this suppose to be a unidirectional or bidirectional mapping?

if bidirectional - then you need to add a reference to actors in `Country` class

if unidirectional - you may skip this comment

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,27 @@
...
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id)
+                .orElseThrow(() -> new DataProcessingException(
+                        String.format("Actor with id=%d was not found", id))
+                );
```
NazarSmith228, line 25: you may inline this
```suggestion
                .orElseThrow(() -> new DataProcessingException("Actor with id=%d was not found".formatted(id)));
```

src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -1,11 +1,28 @@
...
 
+@Entity
+@Table(name = "movies")
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+
+    @ManyToMany
```
NazarSmith228, line 22: the same question as with `Country` class

should this mapping be unidirectional or bidirectional?

NazarSmith228 (changes requested):
Great job! Some minor remarks to fix and some open questions to discuss

src/main/java/mate/academy/hibernate/relations/model/Actor.java
RedArtelerist: I think from the point of business logic, countries don't have to know which actors were born in them, so I will leave this moment unchanged.

src/main/java/mate/academy/hibernate/relations/model/Movie.java
RedArtelerist: I try add bidirectional relational, but tests were failed, so i think i can't add new fields to classes

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
RedArtelerist: Good idea, I'm not sure exactly if we can change AbstactDao, but why not :)

Andriy-dz (changes requested):

src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -1,11 +1,28 @@
...
+    @JoinTable(name = "movie_actors",
```
Andriy-dz, line 23:
```suggestion
    @JoinTable(name = "movies_actors",
```

Andriy-dz (approved):

PR #711, opened 2023-09-25
implemented dao and service level

teract10s:


VolodymyrSkvira (changes requested):
Great solution, just a few issues

Andriy-dz (approved):
Very Good Solution!

PR #719, opened 2023-09-26
added solution for the task

dmytrocherepov:


src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,33 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+        try {
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            session.save(actor);
+            transaction.commit();
+        } catch (RuntimeException e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't add Actor to DB", e);
```
AlexRosenkreuz, line 29: Suggest changing it to make it more informative. You can pass actor name to it, so we'd know which actor causes it. Apply in similar cases

AlexRosenkreuz (approved):
Good job! One minor comment

src/main/java/mate/academy/hibernate/relations/dao/impl/CountryDaoImpl.java
```diff
@@ -12,11 +15,33 @@ public CountryDaoImpl(SessionFactory sessionFactory) {
...
+                session.close();
+            }
+        }
+        return country;
     }
 
     @Override
     public Optional<Country> get(Long id) {
-        return null;
+        Session session = factory.openSession();
```
Elena-Bruyako, line 42: you should use try with resources here

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,34 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+                session.close();
+            }
+        }
+        return actor;
     }
 
     @Override
     public Optional<Actor> get(Long id) {
-        return null;
+        Session session = factory.openSession();
```
Elena-Bruyako, line 42: you should use try with resources here

Elena-Bruyako (changes requested):
Good job! Let's improve your solution!

src/main/java/mate/academy/hibernate/relations/dao/impl/MovieDaoImpl.java
```diff
@@ -12,11 +15,36 @@ public MovieDaoImpl(SessionFactory sessionFactory) {
...
+        }
+        return movie;
     }
 
     @Override
     public Optional<Movie> get(Long id) {
-        return null;
+        try (Session session = factory.openSession()) {
+            Movie movie = session.get(Movie.class, id);
+            session.close();
```
Elena-Bruyako, line 44: it's redundant. You use try with resources here for close the session safety

Elena-Bruyako (approved):

PR #742, opened 2023-10-07
realisation hw hibernate relations

bigvini:


src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,34 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
     @Override
     public Optional<Actor> get(Long id) {
-        return null;
+        Actor actor = null;
+        try (Session session = factory.openSession()) {
+            actor = session.get(Actor.class, id);
+        } catch (Exception e) {
+            throw new DataProcessingException("Can't get actor for id: " + id, e);
+        }
+        return Optional.ofNullable(actor);
```
Ivan95kos, line 46: refactor it everywhere
```suggestion
        try (Session session = factory.openSession()) {
            return Optional.ofNullable(session.get(Actor.class, id));
        } catch (Exception e) {
            throw new DataProcessingException("Can't get actor for id: " + id, e);
        }
```

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,23 @@
...
     @Override
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).get();
```
Ivan95kos, line 21: fix it everywhere
```suggestion
        return actorDao.get(id).orElseThrow(() -> 
                   new EntityNotFoundException("Can't get Actor by id" + id));
```

Ivan95kos (changes requested):
Good job overall!

Ivan95kos (approved):
Nice

PR #745, opened 2023-10-12
solution

kxnxc:


src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,33 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
 
     @Override
     public Actor add(Actor actor) {
-        return null;
+        Session session = null;
```
kerrrusha, line 18: think about utilizing try-with-resouces, it's best practice

src/main/java/mate/academy/hibernate/relations/dao/impl/CountryDaoImpl.java
```diff
@@ -12,11 +15,33 @@ public CountryDaoImpl(SessionFactory sessionFactory) {
 
     @Override
     public Country add(Country country) {
-        return null;
+        Session session = null;
```
kerrrusha, line 18: think about utilizing try-with-resouces, it's best practice

src/main/java/mate/academy/hibernate/relations/dao/impl/MovieDaoImpl.java
```diff
@@ -12,11 +15,33 @@ public MovieDaoImpl(SessionFactory sessionFactory) {
 
     @Override
     public Movie add(Movie movie) {
-        return null;
+        Session session = null;
```
kerrrusha, line 18: think about utilizing try-with-resouces, it's best practice

kerrrusha (changes requested):

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
kxnxc: > think about utilizing try-with-resouces, it's best practice

but in the lectures it was said that for all crud operations except read you should not use try-with-resources since the session can be closed before using transaction.rollback(), should i use TWR anyway?

PR #760, opened 2023-11-06
done

artificialLeo:


pom.xml
```diff
@@ -31,6 +45,12 @@
             <version>2.7.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.hibernate.orm</groupId>
+            <artifactId>hibernate-core</artifactId>
+            <version>6.3.1.Final</version>
+            <scope>compile</scope>
+        </dependency>
```
sokimaaa, line 53: you have 2 dependencies with hibernate... optimize it.

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -2,21 +2,37 @@
...
 public class ActorDaoImpl extends AbstractDao implements ActorDao {
+
     public ActorDaoImpl(SessionFactory sessionFactory) {
         super(sessionFactory);
     }
 
     @Override
     public Actor add(Actor actor) {
-        return null;
+        try (Session session = factory.openSession()) {
```
sokimaaa, line 18: Transaction implemented in wrong way. 
It will not work. 

P.S. Where is your rollback?

src/main/java/mate/academy/hibernate/relations/util/HibernateUtil.java
```diff
@@ -1,9 +1,32 @@
...
 
 import org.hibernate.SessionFactory;
+import org.hibernate.boot.Metadata;
+import org.hibernate.boot.MetadataSources;
+import org.hibernate.boot.registry.StandardServiceRegistry;
+import org.hibernate.boot.registry.StandardServiceRegistryBuilder;
 
 public class HibernateUtil {
+    private static final SessionFactory sessionFactory = buildSessionFactory();
+
```
sokimaaa, line 11: make a private constructor of util class to restrict instantiation of util class.

sokimaaa (changes requested):

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
sokimaaa: Transaction still does not work 

when you call `session.getTransaction` 
session could return diff transactions, so it wont work as you expect.

sokimaaa (changes requested):

kerrrusha (approved):

PR #761, opened 2023-11-07
Hibernate relations hw

dizizzz:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
     @Override
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).orElse(null);
```
kerrrusha, line 23: not critical, but it's better newer to return null from methods
```suggestion
        return actorDao.get(id).orElseThrow(() -> new DataProcessingException("message"));
```

src/main/java/mate/academy/hibernate/relations/service/impl/CountryServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
     @Override
     public Country add(Country country) {
-        return null;
+        return countryDao.add(country);
     }
 
     @Override
     public Country get(Long id) {
-        return null;
+        return countryDao.get(id).orElse(null);
```
kerrrusha, line 23: not critical, but it's better newer to return null from methods
```suggestion
        return countryDao.get(id).orElseThrow(() -> new DataProcessingException("message"));
```

src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
     @Override
     public Movie add(Movie movie) {
-        return null;
+        return movieDao.add(movie);
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        return movieDao.get(id).orElse(null);
```
kerrrusha, line 23: not critical, but it's better newer to return null from methods
```suggestion
        return movieDao.get(id).orElseThrow(() -> new DataProcessingException("message"));
```

src/main/java/mate/academy/hibernate/relations/util/HibernateUtil.java
```diff
@@ -1,9 +1,20 @@
 package mate.academy.hibernate.relations.util;
 
 import org.hibernate.SessionFactory;
+import org.hibernate.cfg.Configuration;
 
 public class HibernateUtil {
+    private static final SessionFactory instance = initSessionFactory();
+
+    private HibernateUtil() {
+        throw new AssertionError("HibernateUtil should not be instantiated.");
```
kerrrusha, line 10: it's definetely not AssertionError. consider choosing of more suitable exception or delete it at all

kerrrusha (changes requested):

src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -2,10 +2,22 @@
...
 
+@Entity
+@Table(name = "movie")
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+
+    @OneToMany
```
Sanderfuck, line 20: I think it can be @ManyToMany and write @JoinTable

Sanderfuck (changes requested):

Sanderfuck (approved):

PR #765, opened 2023-11-09
Solution for task hibernate_relations, first commit

Stelmaschyk:


src/main/java/mate/academy/hibernate/relations/model/Actor.java
```diff
@@ -1,8 +1,20 @@
...
+
+@Entity
+@Table(name = "actors")
 public class Actor implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String name;
+    @OneToOne
     private Country country;
```
olekskov, line 18: I think that should be possible that many actors may be from 1 country

olekskov (approved):
Fine overall, approved.
Please be more careful with relations in future. This time you're lucky that table structure is same for one-to-one and many-to-one relation and there is no logic that may reveal this problem

PR #771, opened 2023-11-13
realized dao and service classes

proxychang:
realized method in classes
annotated model classes and fields

### DIFF
--- a/src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
+++ b/src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
@@ -2,8 +2,11 @@ package mate.academy.hibernate.relations.dao.impl;
 
 import java.util.Optional;
 import mate.academy.hibernate.relations.dao.ActorDao;
+import mate.academy.hibernate.relations.exception.DataProcessingException;
 import mate.academy.hibernate.relations.model.Actor;
+import org.hibernate.Session;
 import org.hibernate.SessionFactory;
+import org.hibernate.Transaction;
 
 public class ActorDaoImpl extends AbstractDao implements ActorDao {
     public ActorDaoImpl(SessionFactory sessionFactory) {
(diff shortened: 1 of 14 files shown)


src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,35 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Data adding failure! "
+                    + "Params: " + actor, ex);
+        } finally {
+            if (session != null) {
+                session.close();
+            }
+        }
```
bhdnchui, line 35: optional: We can reduce code with using try-with-resources here:
```
try (Session session = ...) {
  try {
    transaction = ....
    } catch (Exception e) {
     ...
    }
}
```

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,35 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+        return actor;
     }
 
     @Override
     public Optional<Actor> get(Long id) {
-        return null;
+        try (Session session = factory.openSession()) {
+            Actor actor = session.get(Actor.class, id);
+            return Optional.ofNullable(actor);
```
bhdnchui, line 43:
```suggestion
            return Optional.ofNullable(session.get(Actor.class, id));
```

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id)
+                .orElseThrow(() -> new RuntimeException("Data getting failure."
+                + "Actor with Id: " + id + " not found."));
```
bhdnchui, line 23: It would be better to use String.format here.
Also, let's use narrower exception. IllegalArgumentException for example.
(make sure to do it everywhere)

src/main/java/mate/academy/hibernate/relations/util/HibernateUtil.java
```diff
@@ -1,9 +1,20 @@
...
 
 import org.hibernate.SessionFactory;
+import org.hibernate.cfg.Configuration;
 
 public class HibernateUtil {
+    private static final SessionFactory instance = initSessionFactory();
+
+    private HibernateUtil() {
+
+    }
```
bhdnchui, line 11: Looks like redundant.

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
bhdnchui: (make sure to do it everywhere)

bhdnchui (changes requested):
Good job, just a few points to consider.

bhdnchui (approved):
Good job!

PR #772, opened 2023-11-13
first

Olena-Pro:


src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +16,34 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
 
     @Override
     public Optional<Actor> get(Long id) {
-        return null;
+        try (Session session = factory.openSession()) {
+            return Optional.of(session.get(Actor.class, id));
+        } catch (HibernateException e) {
+            throw new DataProcessingException("Could not get instance of actors", e);
+        } catch (Exception e) {
+            return Optional.empty();
```
olekskov, line 46: Please use Optional.ofNullable to avoid the necessity to of double catch block

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,29 @@
...
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        Optional<Actor> actorOptional = actorDao.get(id);
+        if (actorOptional.isEmpty()) {
+            throw new DataProcessingException("Actor with id = " + id + " no exist.");
+        }
```
olekskov, line 26: 1. Please use orElseThrow method instead of if block
2. Please chose another exception class here. DataProcessingException should mean that we have problem with DB and cannot provide result. In this case everything is fine with DB, we just don't have such record, so NoSuchElementException or EntityNotFoundException will be much more appropriate

olekskov (changes requested):
Please check comment and fix all occurences if multiple in different classes

kozhukhovsky (approved):

PR #791, opened 2023-11-26
practiced in hibernate relations, implemented dao classes, hibernate …

maxymmusiienko:
…util


src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,23 @@
...
     @Override
     public Movie add(Movie movie) {
-        return null;
+        return movieDao.add(movie);
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        return movieDao.get(id).orElse(null);
```
okuzan, line 21: orElseThrow EntityNotFoundException with message that includes id. fix everywhere

okuzan (changes requested):

okuzan (approved):
Well done!

PR #795, opened 2023-11-29
first solution

IhorTrokhymchuk:


src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,23 @@
...
     @Override
     public Movie add(Movie movie) {
-        return null;
+        return movieDao.add(movie);
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        return movieDao.get(id).orElse(null);
```
olekskov, line 21: Returning null from Optional nearly makes no sense.
Better to throw relevant exception with proper message

```suggestion
        return movieDao.get(id).orElseThrow(new EntityNotFoundException("Movie with id " + id + " is not found));
```

olekskov (approved):

PR #807, opened 2023-12-24
solved task

AntonKurylo:


src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -2,10 +2,26 @@
...
 
+@Entity
+@Table(name = "movies")
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+    @OneToMany
+    @JoinTable(name = "movies_actors",
```
Sanderfuck, line 22: Are you sure about this relation?

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
+import mate.academy.hibernate.relations.dao.impl.ActorDaoImpl;
 import mate.academy.hibernate.relations.model.Actor;
 import mate.academy.hibernate.relations.service.ActorService;
+import org.hibernate.SessionFactory;
 
 public class ActorServiceImpl implements ActorService {
+    private final ActorDao actorDao;
+
+    public ActorServiceImpl(SessionFactory sessionFactory) {
+        this.actorDao = new ActorDaoImpl(sessionFactory);
```
Sanderfuck, line 13: Don`t init actor dao in service layer, do it in Dao layer

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
     @Override
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).orElse(null);
```
Sanderfuck, line 23: orElse null is very bad idea, use orElseThrow

Sanderfuck (changes requested):

src/main/java/mate/academy/hibernate/relations/model/Movie.java
AntonKurylo: Why not? There can be several actors in one movie

PR #820, opened 2024-01-11
solution 11/01

VadimKadasiev:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
     @Override
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).orElse(null);
```
okuzan, line 23: orElseThrow EntityNotFoundException with message + id, fix everywhere

src/main/java/mate/academy/hibernate/relations/util/HibernateUtil.java
```diff
@@ -1,9 +1,19 @@
 package mate.academy.hibernate.relations.util;
 
 import org.hibernate.SessionFactory;
+import org.hibernate.cfg.Configuration;
 
 public class HibernateUtil {
+    private static final SessionFactory INSTANCE = createSessionFactoryInstance();
+
+    public HibernateUtil() {
```
okuzan, line 9:
```suggestion
    private HibernateUtil() {
```

okuzan (changes requested):

okuzan (approved):

PR #825, opened 2024-01-15
Implement DAO, Service layers and Models

MurphyEdward:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,24 @@
...
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        Optional<Actor> optional = actorDao.get(id);
+
+        return optional.orElse(null);
```
olekskov, line 22: By returning null instead of optional you makes it meaningless on dao layer.
We should provide default value or throw suitable exception
```suggestion
        return actorDao.get(id).orElseThrow(() -> new EntityNotFoundException("actor with id...."));
```

src/main/java/mate/academy/hibernate/relations/service/impl/CountryServiceImpl.java
```diff
@@ -1,16 +1,24 @@
...
-        return null;
+        return countryDao.add(country);
     }
 
     @Override
     public Country get(Long id) {
-        return null;
+        Optional<Country> optional = countryDao.get(id);
+
+        return optional.orElse(null);
```
olekskov, line 22: same

src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,24 @@
...
-        return null;
+        return movieDao.add(movie);
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        Optional<Movie> optional = movieDao.get(id);
+
+        return optional.orElse(null);
```
olekskov, line 22: same

olekskov (changes requested):

kozhukhovsky (approved):

PR #834, opened 2024-02-02
completed implementation model, dao/impl, service/impl/ created DataP…

RCherniavskiy:
…rocessingException, added HibernateUtil

### DIFF
--- a/src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
+++ b/src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
@@ -1,16 +1,27 @@
 package mate.academy.hibernate.relations.service.impl;
 
+import mate.academy.hibernate.relations.dao.ActorDao;
+import mate.academy.hibernate.relations.exception.DataProcessingException;
 import mate.academy.hibernate.relations.model.Actor;
 import mate.academy.hibernate.relations.service.ActorService;
 
 public class ActorServiceImpl implements ActorService {
+    private final ActorDao actorDao;
+
+    public ActorServiceImpl(ActorDao actorDao) {
+        this.actorDao = actorDao;
+    }
     @Override
     public Actor add(Actor actor) {
-        return null;
+        try {
+            return actorDao.add(actor);
+        } catch (Exception e) {
+            throw new DataProcessingException("Failed to add to the Db actor: " + actor);
+        }
     }
-
     @Override
-    public Actor get(Long id) {
-        return null;
+    public Actor get(Long id){
+        return actorDao.get(id)
+                .orElseThrow(() -> new DataProcessingException("Actor not found with id: " + id));
     }
 }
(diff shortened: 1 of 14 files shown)


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,29 @@
...
+    }
+
     @Override
     public Actor add(Actor actor) {
-        return null;
+        try {
+            return actorDao.add(actor);
+        } catch (Exception e) {
+            throw new DataProcessingException("Failed to add to the Db actor: " + actor);
+        }
```
olekskov, line 21: Your dao layer will throw exception if something is wrong, why we need try/catch here?
```suggestion
        return actorDao.add(actor);
```

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,29 @@
...
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id)
+                .orElseThrow(() -> new DataProcessingException("Actor not found with id: " + id));
```
olekskov, line 27: Better to use different exceptions when there are problems with DB (DataProcessing) and when everything is ok and we just received request to find something we don't have in DB (EntityNotFound)
```suggestion
        return actorDao.get(id)
                .orElseThrow(() -> new EntityNotFoundException("Actor not found with id: " + id));
```

src/main/java/mate/academy/hibernate/relations/service/impl/CountryServiceImpl.java
```diff
@@ -1,16 +1,29 @@
...
+        } catch (Exception e) {
+            throw new DataProcessingException("Failed to add to the Db actor: " + country);
+        }
     }
 
     @Override
     public Country get(Long id) {
-        return null;
+        return countryDao.get(id)
+                .orElseThrow(() -> new DataProcessingException("Actor not found with id: " + id));
```
olekskov, line 27: same points

src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,29 @@
...
+            throw new DataProcessingException("Failed to add to the Db actor: " + movie);
+        }
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        return movieDao.get(id)
+                .orElseThrow(() -> new DataProcessingException("Actor not found with id: " + id));
     }
```
olekskov, line 28: same

olekskov (changes requested):

olekskov (approved):
Good job

PR #838, opened 2024-02-07
task solution mark.1

andriimikhaylovskiy:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,29 @@
...
     @Override
     public Actor get(Long id) {
-        return null;
+        Optional<Actor> actor = dao.get(id);
+        if (actor.isPresent()) {
+            return actor.get();
+        } else {
+            throw new RuntimeException("No actor with id=" + id + "is found.");
+        }
     }
```
okuzan, line 28: instead of all of this just call orElseThrow + enitytnotfound +message + id

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,15 @@
...
+        <property name="connection.username">root</property>
+        <property name="connection.password">1234</property>
+        <property name="show_sql">true</property>
+        <property name="hbm2ddl.auto">create-drop</property>
+
+        <mapping class="mate.academy.hibernate.relations.model.Movie"/>
+        <mapping class="mate.academy.hibernate.relations.model.Actor"/>
+        <mapping class="mate.academy.hibernate.relations.model.Country"/>
+    </session-factory>
+</hibernate-configuration>
```
okuzan, line 15: if you don't end ur files with 1 empty line, u will see this symbol
image

okuzan (changes requested):
fix get method in each service

olekskov (approved):

PR #843, opened 2024-02-13
all done

dima-holovko:


src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,28 @@
...
-        return null;
+
+        return movieDao.add(movie);
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        Optional<Movie> optionalMovie = movieDao.get(id);
+        return optionalMovie.get();
```
okuzan, line 26: just `get()` on optional is a bad practice, use orElseThrow + entityNotFound + message + id

okuzan (changes requested):

okuzan (approved):

PR #850, opened 2024-02-23
HW relations

Chorniypistolet:


src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -2,21 +2,57 @@
...
+        Actor actor;
+        try {
+            actor = session.get(Actor.class, id);
+        } catch (HibernateException e) {
+            throw new DataProcessingException("Error getting actor from db", e);
+        } finally {
+            if (session != null) {
+                session.close();
+            }
+        }
```
olekskov, line 55: for get operations we may use try-with-resources to simplify code

src/main/java/mate/academy/hibernate/relations/model/Actor.java
```diff
@@ -1,8 +1,18 @@
...
+import jakarta.persistence.OneToOne;
+
+@Entity
 public class Actor implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String name;
+    @OneToOne
     private Country country;
```
olekskov, line 16: I don't think that there is only one actor in each country

src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -1,11 +1,20 @@
...
 import java.util.List;
 
+@Entity
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+    @OneToMany
     private List<Actor> actors;
```
olekskov, line 18: And don't think that actor can be only in one movie, it would be very short career

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,24 @@
...
     @Override
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).orElse(null);
```
olekskov, line 22: By returning null from service we're making Optional usage on dao layer totally useless
```suggestion
        return actorDao.get(id).orElseThrow(() -> new EntityNotFoundException("with proper message"));
```
Please recheck all similar cases

olekskov (changes requested):

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -2,21 +2,52 @@
...
+import org.hibernate.Transaction;
 
 public class ActorDaoImpl extends AbstractDao implements ActorDao {
+
+    private final SessionFactory sessionFactory;
+
     public ActorDaoImpl(SessionFactory sessionFactory) {
         super(sessionFactory);
+        this.sessionFactory = sessionFactory;
     }
```
Rommelua, line 19:
```suggestion
public class ActorDaoImpl extends AbstractDao implements ActorDao {
    public ActorDaoImpl(SessionFactory sessionFactory) {
        super(sessionFactory);
    }
```

Rommelua (approved):

PR #856, opened 2024-03-14
feat: add solution

breakingtumble:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,27 @@
...
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).orElseThrow(
+                () -> new RuntimeException(
```
olekskov, line 24: We should use more precise exception types.
Futher they will be converted to proper response to user and this conversion will be much more simple if we have proper exception types
```suggestion
                () -> new EntityNotFoundException(
```

src/main/java/mate/academy/hibernate/relations/service/impl/CountryServiceImpl.java
```diff
@@ -1,16 +1,27 @@
...
     public Country add(Country country) {
-        return null;
+        return countryDao.add(country);
     }
 
     @Override
     public Country get(Long id) {
-        return null;
+        return countryDao.get(id).orElseThrow(
+                () -> new RuntimeException(
```
olekskov, line 24: same

src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,27 @@
...
     public Movie add(Movie movie) {
-        return null;
+        return movieDao.add(movie);
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        return movieDao.get(id).orElseThrow(
+                () -> new RuntimeException(
```
olekskov, line 24: same

olekskov (changes requested):

olekskov (approved):

PR #863, opened 2024-03-21
task done

AndrewOliferchuk:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,27 @@
...
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        if (actorDao.get(id).isEmpty()) {
+            throw new DataProcessingException("Can't find actor by id: " + id);
+        }
+        return actorDao.get(id).get();
```
olekskov, line 25: 1. Use orElseThrow method to make code shorter
2. DataProcessingException is used on dao layer to indicate problems in communication with DB (situation where result cannot be determined). In service another type of exception should be used as operation with DB was successful and we can properly reply that we don't have requested data
```suggestion
        return actorDao.get(id)
                .orElseThrow(() -> new EntityNotFoundException("Can't find actor by id: " + id));
```

src/main/java/mate/academy/hibernate/relations/service/impl/CountryServiceImpl.java
```diff
@@ -1,16 +1,27 @@
...
+        return countryDao.add(country);
     }
 
     @Override
     public Country get(Long id) {
-        return null;
+        if (countryDao.get(id).isEmpty()) {
+            throw new DataProcessingException("Can't find country by id: " + id);
+        }
+        return countryDao.get(id).get();
```
olekskov, line 25: same

src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,27 @@
...
+        return movieDao.add(movie);
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        if (movieDao.get(id).isEmpty()) {
+            throw new DataProcessingException("Can't find movie by id: " + id);
+        }
+        return movieDao.get(id).get();
```
olekskov, line 25: same

olekskov (changes requested):

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,15 @@
...
+        <property name="connection.username">root</property>
+        <property name="connection.password">Password</property>
+        <property name="show_sql">true</property>
+        <property name="hbm2ddl.auto">create-drop</property>
+
+        <mapping class="mate.academy.hibernate.relations.model.Actor"/>
+        <mapping class="mate.academy.hibernate.relations.model.Country"/>
+        <mapping class="mate.academy.hibernate.relations.model.Movie"/>
+    </session-factory>
+</hibernate-configuration>
```
okuzan, line 15:
```suggestion
</hibernate-configuration>

```

src/main/java/mate/academy/hibernate/relations/util/HibernateUtil.java
```diff
@@ -1,9 +1,19 @@
 package mate.academy.hibernate.relations.util;
 
 import org.hibernate.SessionFactory;
+import org.hibernate.cfg.Configuration;
 
 public class HibernateUtil {
+    private static final SessionFactory instance = initSessionFactory();
+
+    public HibernateUtil() {
```
okuzan, line 9: utility classes should have private constructors to prevent instantiation

okuzan (changes requested):

olekskov (approved):

PR #869, opened 2024-03-27
completed implementation models,daoImpl,serviceImpl

DmytroHadiuchko:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        Optional<Actor> actorList = actorDao.get(id);
+        return actorList.orElse(null);
```
makaroshyna, line 23: Here it would be better to use or else throw and throw our custom exception.
Do it in other services too.

makaroshyna (changes requested):
Nice! Left some comments)

PR #871, opened 2024-03-29
main version

EduarfF:


src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,32 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+        try {
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            session.persist(actor);
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't add actor: " + actor, e);
```
VlasenkoYevhenii, line 29: I'd rather create constants for the messages. But it's up to you here

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,31 @@
...
+    public ActorServiceImpl(ActorDao actorDao) {
+        this.actorDao = actorDao;
+    }
+
     @Override
     public Actor add(Actor actor) {
-        return null;
+        if (actor == null) {
+            throw new DataProcessingException("Cannot add null data");
+        }
```
VlasenkoYevhenii, line 19: I think this will be caught by Dao layer and is not needed here

src/main/java/mate/academy/hibernate/relations/service/impl/CountryServiceImpl.java
```diff
@@ -1,16 +1,31 @@
...
+        return country;
     }
 
     @Override
     public Country get(Long id) {
-        return null;
+        if (id == null) {
+            throw new DataProcessingException("Cannot get by null id");
+        }
+        return countryDao.get(id).orElseThrow(DataProcessingException::new);
```
VlasenkoYevhenii, line 29: I believe the custom exception is supposed to be thrown on Dao layer only. Here it would be better to throw NoSuchElementException with some message

VlasenkoYevhenii (changes requested):
GJ!
Check the comments pls!

andrii-hoienko (approved):
Good job!

PR #876, opened 2024-03-31
Added implementation for task hibernate relations.

nklimovych:


src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,38 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
 
     @Override
     public Actor add(Actor actor) {
-        return null;
+        Session session = null;
+        Transaction transaction = null;
+        try {
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            try {
```
slizko1, line 23: Is second try-catch block necessary?

slizko1 (approved):
Seems nice to me except second try-catch bloks in DaoImpl classes

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
nklimovych: It is not necessary, but I added it to catch specific exceptions to separate two different situations when you can't add something to the database or just unable to open a session.

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,27 @@
...
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        if (actorDao.get(id).isEmpty()) {
+            throw new DataProcessingException("There's no actor with specified id: " + id);
+        }
+        return actorDao.get(id).get();
```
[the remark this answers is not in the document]
nklimovych: As I understand, we cannot use orElseThrow() in this case as we are returning an Actor instead of Optional.

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,33 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            session.persist(actor);
+            transaction.commit();
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException(
+                    "Can't add actor to database, id = " + actor.getId(), e);
```
andrii-hoienko, line 30: Actor will not have id when it is fully new, because id will be setup by hibernate. I suggest just add fill actor abject to response

andrii-hoienko: Check similar cases

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,15 @@
...
+        <property name="connection.username">root</property>
+        <property name="connection.password">password</property>
+        <property name="show_sql">true</property>
+        <property name="hbm2ddl.auto">create-drop</property>
+
+        <mapping class="mate.academy.hibernate.relations.model.Actor"/>
+        <mapping class="mate.academy.hibernate.relations.model.Country"/>
+        <mapping class="mate.academy.hibernate.relations.model.Movie"/>
+    </session-factory>
+</hibernate-configuration>
```
andrii-hoienko, line 15:
```suggestion
</hibernate-configuration>

```

andrii-hoienko (changes requested):
Very nice, left minor comments

pom.xml
```diff
@@ -31,6 +31,16 @@
             <version>2.7.2</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>mysql</groupId>
+            <artifactId>mysql-connector-java</artifactId>
+            <version>8.0.33</version>
+        </dependency>
```
karma-o, line 38: I am not sure we need this one

karma-o (approved):

PR #885, opened 2024-04-01
Created hibernate relations hw realisation

Recdt:


src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -2,10 +2,19 @@
...
+import javax.persistence.Id;
+import javax.persistence.OneToMany;
 
+@Entity
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+    @OneToMany
```
fedaktaras, line 17: I think ManyToMany is better. But as it passed tests it might be ok.

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,50 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't add an actor to DB "
+                    + actor.toString(), e);
+        } finally {
+            if (session != null) {
+                session.close();
+            }
```
fedaktaras, line 37: As I know openSesion can't return null it can throw HibernateException which is RuntimeException

fedaktaras: But i beleive it from the lesson. So I have to check it. Might be good.

fedaktaras (approved):
I have some questions but I am not sure to reject. I might be wrong so I approve.

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
karma-o: I guess what you were trying to say is that we catch usual `Exception` here and not `Runtime`? if yes - `RuntimeException` is an instance of `Exception`

src/main/java/mate/academy/hibernate/relations/model/Movie.java
karma-o: Agree, one actor can act in many Movies, and one movie has multiple actors, so that is a many-to-may relation

src/main/java/mate/academy/hibernate/relations/model/Actor.java
```diff
@@ -1,8 +1,18 @@
...
+import javax.persistence.Id;
+import javax.persistence.OneToOne;
+
+@Entity
 public class Actor implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String name;
+    @OneToOne
```
karma-o, line 15: One country can have a maximum of 1 actor? I guess their salaries must be huge if we have only around 200 of them 😄 
How many actors can 1 county have?
How many home counties can 1 actor have?
What would be a relation between them, judjing by the answers to those 2 last questions? always ask them yourself if you are not sure.

We also need to specify the `JoinColumn`/`JoinTable` depending on what relation you have, that goes for other entities as well

src/main/resources/hibernate.cfg.xml
```diff
@@ -0,0 +1,15 @@
...
+        <property name="connection.username">root</property>
+        <property name="connection.password">55555Qwz!</property>
+        <property name="show_sql">true</property>
+        <property name="hbm2ddl.auto">create-drop</property>
+
+        <mapping class="mate.academy.hibernate.relations.model.Actor"/>
+        <mapping class="mate.academy.hibernate.relations.model.Movie"/>
+        <mapping class="mate.academy.hibernate.relations.model.Country"/>
+    </session-factory>
+</hibernate-configuration>
```
karma-o, line 15:
```suggestion
</hibernate-configuration>

```

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,35 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
...
+            session.save(actor);
+            transaction.commit();
+
+            return actor;
+        } catch (Exception e) {
+            if (transaction != null) {
+                transaction.rollback();
+            }
+            throw new DataProcessingException("Can't add an actor to DB "
+                    + actor.toString(), e);
```
karma-o, line 33: `toString()` call is redundant, also I'd rather not log personal information of an actor

karma-o (changes requested):
Refresh your knowledge on the type of the relations between different entities

src/main/java/mate/academy/hibernate/relations/model/Actor.java
```diff
@@ -1,8 +1,18 @@
 package mate.academy.hibernate.relations.model;
 
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.persistence.OneToOne;
+
+@Entity
 public class Actor implements Cloneable {
```
karma-o, line 10: add `@Table` for each entity

src/main/java/mate/academy/hibernate/relations/model/Actor.java
Recdt: I explain this for my self like a one actor can have only one country, but thx for hint

PR #904, opened 2024-05-01
implemented models, dao and services

nastasiiaKulinich:


src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,23 @@
...
     @Override
     public Movie add(Movie movie) {
-        return null;
+        return movieDao.add(movie);
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        return movieDao.get(id).get();
```
okuzan, line 21: you are losing the benefit of having optional in the first place - you can get NPE here. Use orElseThrow + entity not found exception + message + id, everywhere

okuzan (changes requested):

src/main/java/mate/academy/hibernate/relations/dao/impl/ActorDaoImpl.java
```diff
@@ -12,11 +15,32 @@ public ActorDaoImpl(SessionFactory sessionFactory) {
 
     @Override
     public Actor add(Actor actor) {
-        return null;
+        Session session = null;
+        Transaction transaction = null;
+        try {
+            session = factory.openSession();
+            transaction = session.beginTransaction();
+            session.save(actor);
```
Elena-Bruyako, line 23:
```suggestion
            session.persist(actor);
```
Check other implementation

src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -1,11 +1,28 @@
...
 import java.util.List;
 
+@Entity
+@Table(name = "movies")
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+    @ManyToMany(fetch = FetchType.EAGER)
```
Elena-Bruyako, line 22:
```suggestion
    @ManyToMany
```

Elena-Bruyako (changes requested):

PR #911, opened 2024-05-10
implemented relations between entities

vladshender:


src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -2,10 +2,22 @@
...
 
+@Entity
+@Table(name = "movies")
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+
+    @OneToMany
```
okuzan, line 20: actor can act only in one movie?

okuzan (changes requested):

okuzan (approved):
Good job

PR #925, opened 2024-06-01
Hw hrelations

Hryniv:


pom.xml
```diff
@@ -31,7 +31,18 @@
             <version>2.7.2</version>
```
Elena-Bruyako, line 31: Didn't find the Main class where you implement // TODO:

Elena-Bruyako (changes requested):

src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,25 @@
...
     public Movie add(Movie movie) {
-        return null;
+        return movieDao.add(movie);
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        return movieDao.get(id).orElseThrow(() ->
+                new DataProcessingException("Can't get movie by id " + id));
```
okuzan, line 23: use EntityNotFoundException for service layer, DataProcessingException is more suitable for DAO layer

okuzan (changes requested):

Elena-Bruyako (approved):

PR #937, opened 2024-07-10
added solution of task

apanyutin:


src/main/java/mate/academy/hibernate/relations/Main.java
```diff
@@ -7,26 +7,29 @@
...
 
 public class Main {
     public static void main(String[] args) {
         // use this session factory when you will initialize service instances
-        SessionFactory sessionFactory = HibernateUtil.getSessionFactory();
+        SessionFactory sessionFactory = HibernateUtil.getInstance();
 
         Country usa = new Country("USA");
-        CountryService countryService = null; // TODO: initialize this instance
+        CountryService countryService = new CountryServiceImpl(sessionFactory);
```
Elena-Bruyako, line 22:
```suggestion
        CountryService countryService = new CountryServiceImpl(new CountryDaoImpl(sessionFactory));
```

same for other services

Elena-Bruyako, line 22:
```suggestion
        CountryService countryService = new CountryServiceImpl(new CountryDaoImpl(sessionFactory));
```

same for other services

src/main/java/mate/academy/hibernate/relations/model/Actor.java
```diff
@@ -1,8 +1,20 @@
...
+import jakarta.persistence.Table;
+
+@Entity
+@Table (name = "actors")
 public class Actor implements Cloneable {
+    @Id
+    @GeneratedValue (strategy = GenerationType.IDENTITY)
     private Long id;
     private String name;
+    @ManyToOne
```
Elena-Bruyako, line 17:
```suggestion
    @ManyToOne
    @JoinColumn....
```

src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -1,11 +1,22 @@
...
 import java.util.List;
 
+@Entity
+@Table(name = "movies")
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+    @ManyToMany
```
Elena-Bruyako, line 19: Add `@JoinTable.....`

src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,27 @@
...
     public Actor add(Actor actor) {
-        return null;
+        return actorDao.add(actor);
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        return actorDao.get(id).orElseThrow(
+                () -> new RuntimeException("DB don't have actor with such id = " + id));
```
Elena-Bruyako, line 25:
```suggestion
                () -> new EntityNotFoundException("DB don't have actor with such id = " + id));
```

src/main/java/mate/academy/hibernate/relations/service/impl/CountryServiceImpl.java
```diff
@@ -1,16 +1,27 @@
...
     public Country add(Country country) {
-        return null;
+        return countryDao.add(country);
     }
 
     @Override
     public Country get(Long id) {
-        return null;
+        return countryDao.get(id).orElseThrow(
+                () -> new RuntimeException("DB don't have country with such id = " + id));
```
Elena-Bruyako, line 25: the same

src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,27 @@
...
     public Movie add(Movie movie) {
-        return null;
+        return movieDao.add(movie);
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        return movieDao.get(id).orElseThrow(
+                () -> new RuntimeException("DB don't have movie with such id = " + id));
```
Elena-Bruyako, line 25: the same

Elena-Bruyako (changes requested):

src/main/java/mate/academy/hibernate/relations/Main.java
apanyutin: Why need to do this? Why my option is not correct?

apanyutin: Why need I to do this? Why my option is not correct?

src/main/java/mate/academy/hibernate/relations/model/Actor.java
apanyutin: Why need I add @JoinColumn in this case?

src/main/java/mate/academy/hibernate/relations/model/Movie.java
apanyutin: Why need I add @JoinColumn in this case?

src/main/java/mate/academy/hibernate/relations/Main.java
Elena-Bruyako: You should provide isolation and abstraction and dep injection
Why do you need to make such logic in service?


```
public ActorServiceImpl(SessionFactory sessionFactory) {
        actorDao = new ActorDaoImpl(HibernateUtil.getInstance());
    }
```

src/main/java/mate/academy/hibernate/relations/model/Movie.java
Elena-Bruyako: Note that using @JoinTable or even @JoinColumn isn’t required. JPA will generate the table and column names for us. However, the strategy JPA uses won’t always match the naming conventions we use. So, we need the possibility to configure table and column names.

src/main/java/mate/academy/hibernate/relations/model/Actor.java
Elena-Bruyako: Note that using @JoinTable or even @JoinColumn isn’t required. JPA will generate the table and column names for us. However, the strategy JPA uses won’t always match the naming conventions we use. So, we need the possibility to configure table and column names.

Elena-Bruyako (changes requested):

src/main/java/mate/academy/hibernate/relations/model/Movie.java
apanyutin: Please look at my solution. I don't sure of column naming, because I dont have deep understanding what is joinColumns and inverseJoinColumns.

MentorMateAcademy (approved):

PR #941, opened 2024-07-16
Hibernate hw

Yaromyros:
Solved homework with relations in hibernate


src/main/java/mate/academy/hibernate/relations/service/impl/MovieServiceImpl.java
```diff
@@ -1,16 +1,26 @@
...
-        return null;
+        movieDao.add(movie);
+        return movie;
     }
 
     @Override
     public Movie get(Long id) {
-        return null;
+        return movieDao.get(id).orElseThrow(() ->
+                new DataProcessingException("Unable to get movie with id " + id));
```
okuzan, line 24: for Service layer, please use EntityNotFoundException instead

okuzan (changes requested):

Yaromyros:
Thanks, did not know))

MentorMateAcademy (approved):

PR #950, opened 2024-08-01
Initial commit: hibernate relations hw

OleksiiLytvynenko:


src/main/java/mate/academy/hibernate/relations/service/impl/CountryServiceImpl.java
```diff
@@ -1,16 +1,23 @@
...
     @Override
     public Country add(Country country) {
-        return null;
+        return countryDao.add(country);
     }
 
     @Override
     public Country get(Long id) {
-        return null;
+        return countryDao.get(id).orElseThrow();
```
okuzan, line 21: fix everywhere - this will throw generic nosuchelementexception, let's throw more specific EntityNotFoundException + message + id

okuzan (changes requested):

okuzan (approved):

PR #1013, opened 2024-11-26
implemented relations between entities and realized methods

Nazar1106:


src/main/java/mate/academy/hibernate/relations/service/impl/CountryServiceImpl.java
```diff
@@ -1,16 +1,23 @@
...
     @Override
     public Country add(Country country) {
-        return null;
+        return countryDao.add(country);
     }
 
     @Override
     public Country get(Long id) {
-        return null;
+        return countryDao.get(id).orElseThrow();
```
okuzan, line 21: throw entity not found +  message + id everywhere

src/main/java/mate/academy/hibernate/relations/model/Movie.java
```diff
@@ -1,11 +1,20 @@
...
 import java.util.ArrayList;
 import java.util.List;
 
+@Entity
 public class Movie implements Cloneable {
+    @Id
+    @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String title;
+    @ManyToMany
```
okuzan, line 17: add JoinTable

okuzan (changes requested):

PR #1098, opened 2025-06-02
hibernate relations

DariaBakal:


DariaBakal:
@mateacademy-ai-mentor pom.xml file was added to the PR in the first commit

PR #1118, opened 2025-08-24
Implement Hibernate relations homework

Daniul2:
## Summary

Implements the Hibernate relations homework: model entities, DAO layer, service layer, custom exception, Hibernate configuration and `HibernateUtil`, with a runnable `Main`.

- `Country` (one-to-many), `Actor` (many-to-one with `Country`), `Movie` (many-to-many with `Actor`)
- DAO impls extend `AbstractDao` (constructor-injected `SessionFactory`) and throw a custom unchecked `DataProcessingException`
- `hibernate.cfg.xml` + `HibernateUtil` singleton `SessionFactory`

## Key fix

`CountryDaoImpl.add()` / `ActorDaoImpl.add()` switched from `session.merge()` to `session.persist()`. `merge()` returns a new managed copy but leaves the original object's id as null, so tests reusing the original reference (e.g. usaClone as an actor's country) hit `TransientObjectException`. `persist()` assigns the generated id directly to the passed-in object. `MovieDaoImpl.add()` keeps `merge()` to also support updating an existing movie's actor list.

## Test plan

- [x] `mvn clean test` - all 11 DAO tests pass (HSQLDB in-memory)
- [x] `Main` runs end-to-end against MySQL without errors

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -11,49 +11,36 @@
     <properties>
         <jdk.version>17</jdk.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <maven.checkstyle.plugin.version>3.1.1</maven.checkstyle.plugin.version>
-        <maven.checkstyle.plugin.configLocation>
-            https://raw.githubusercontent.com/mate-academy/style-guides/master/java/checkstyle.xml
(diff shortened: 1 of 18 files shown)


Daniul2:
@mateacademy-ai-mentor please re-review, I've added the private constructor to HibernateUtil

Daniul2:
@mateacademy-ai-mentor please re-review, all requested changes have been fixed

PR #1159, opened 2026-02-03
Hw hib relations

sannchous:


src/main/java/mate/academy/hibernate/relations/service/impl/ActorServiceImpl.java
```diff
@@ -1,16 +1,30 @@
...
     }
 
     @Override
     public Actor get(Long id) {
-        return null;
+        if (id < 0) {
+            throw new IllegalArgumentException("Id cannot be negative");
+        }
+        return actorDao.get(id).orElseThrow(
+                () -> new RuntimeException("Actor Not Found, id = " + id));
```
Elena-Bruyako, line 28: use `EntityNotFoundException` in service layer
```suggestion
        if (id < 0) {
            throw new IllegalArgumentException("Id cannot be negative");
        }
        return actorDao.get(id).orElseThrow(
                () -> new EntityNotFoundException("Actor Not Found, id = " + id));
```

Elena-Bruyako (approved):
Good job! 
just one comment for ther future

